PR #1, opened 2023-02-28
Initial setup and addition of files.

yadavshashankr:


app/build.gradle
```diff
@@ -4,12 +4,12 @@ plugins {
...
 
 android {
-    compileSdk 33
+    compileSdk 32
 
     defaultConfig {
         applicationId "com.portfolio.tasky"
         minSdk 30
-        targetSdk 33
+        targetSdk 32
```
MKiperszmid, line 12: Why are you targetting 32 and not 33?

yadavshashankr: The IDE caches are causing some problem when targeting to 33. I'll make sure they are reset and in the next PR I'll target it to 33.

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    override fun onResume() {
+        super.onResume()
+        // Not implemented yet
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        // Not implemented yet
+    }
```
MKiperszmid, line 43: No need to add this on this PR if they aren't being used right now

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
```diff
@@ -0,0 +1,50 @@
...
+import android.content.Intent
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+
+open class BaseActivity : AppCompatActivity() {
+    private var mActivity: BaseActivity? = null
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        mActivity = this
```
MKiperszmid, line 12: Why would you have an instance of yourself?

I know it's for later. But im curious

app/src/main/java/com/portfolio/tasky/BaseFragment.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    override fun onStop() {
+        super.onStop()
+        // Not implemented yet
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        // Not implemented yet
+    }
```
MKiperszmid, line 37: Same here. No need to add this lines of code if they aren't being used in this pr

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
```diff
@@ -0,0 +1,50 @@
+package com.portfolio.tasky
+
+import android.content.Intent
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+
+open class BaseActivity : AppCompatActivity() {
```
MKiperszmid, line 7: I don't recommend having Base, unless it's something ultra specific

What's your intention for using BaseActivity and BaseFragment?

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+import android.view.View
+import android.view.ViewGroup
+import androidx.transition.Slide
+import com.portfolio.tasky.BaseFragment
+import com.portfolio.tasky.databinding.LayoutLoginBinding
+
+
+
+private const val ARG_PARAM1 = "param1"
+private const val ARG_PARAM2 = "param2"
```
MKiperszmid, line 16: Most stuff that's not being used, could be removed and added later :D

app/src/main/res/values-night/themes.xml
```diff
@@ -10,7 +10,9 @@
         <item name="colorSecondaryVariant">@color/teal_200</item>
         <item name="colorOnSecondary">@color/black</item>
         <!-- Status bar color. -->
-        <item name="android:statusBarColor" tools:targetApi="l">?attr/colorPrimaryVariant</item>
+        <item name="android:statusBarColor" >?attr/colorPrimaryVariant</item>
```
MKiperszmid, line 13: There's an extra space between
"android:statusBarColor"
and
\>?attr/colorPrimaryVariant

app/src/main/res/values/themes.xml
```diff
@@ -1,16 +1,17 @@
...
+        <item name="colorPrimary">@color/black</item>
         <item name="colorPrimaryVariant">@color/purple_700</item>
         <item name="colorOnPrimary">@color/white</item>
         <!-- Secondary brand color. -->
         <item name="colorSecondary">@color/teal_200</item>
         <item name="colorSecondaryVariant">@color/teal_700</item>
         <item name="colorOnSecondary">@color/black</item>
         <!-- Status bar color. -->
-        <item name="android:statusBarColor" tools:targetApi="l">?attr/colorPrimaryVariant</item>
+        <item name="android:statusBarColor" >?attr/colorPrimaryVariant</item>
```
MKiperszmid, line 13: There's an extra space here too

MKiperszmid (approved):
Some minor stuff

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
yadavshashankr: Since the BaseActivity and BaseFragment will serve as an extension to all other activities/fragments, it will be easy for me to access activity's context which is in foreground easily. (As per my understandings)

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
yadavshashankr: Sure, thanks for the videos!

BaseActivity and BaseFragment will accumulate all the logic to control generic components like Toolbars, logging for app traversals, independent dialogs etc.

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt

app/src/main/java/com/portfolio/tasky/BaseFragment.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
yadavshashankr: Done

app/src/main/res/values/themes.xml
yadavshashankr: Done

app/src/main/res/values-night/themes.xml
yadavshashankr: Done

app/build.gradle
MKiperszmid: Sounds good!

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
MKiperszmid: I recommend not doing that, instead have Util classes or helpers that handle that logic
Charfaoui makes a good video about it and how you can do it: https://www.youtube.com/watch?v=YtbwxEL4uy8

yadavshashankr: Okay, Thanks!

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
philipplackner: I'd recommend not to use any base classes (at least if they're called `BaseX`), since that really smells like a violation of the single responsibility principle. If there's one clear responsibility for the class, you could give it a clear name which `BaseActivity` or `BaseFragment` isn't. 
Having base classes couples all classes to that base class, so any change in the base class will have an impact on all other classes which you can usually find better solutions for. What's reasonable is to have a root fragment for example for the ViewBinding setup, but then this root fragment has a clear responsibility and could be called `BindingFragment` or similar

app/src/main/java/com/portfolio/tasky/BaseActivity.kt
yadavshashankr: Done! Removed BaseActivity and BaseFragment from the project.

app/src/main/AndroidManifest.xml
```diff
@@ -12,6 +12,19 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/Theme.Tasky"
-        tools:targetApi="31" />
+        tools:targetApi="31">
+        <activity
+            android:name=".BaseActivity"
+            android:exported="false" />
```
MKiperszmid, line 18: Since we removed base activity, we should remove it from here also

MKiperszmid (approved):

app/src/main/AndroidManifest.xml
yadavshashankr: Terrible mistake !! 

Done

PR #2, opened 2023-03-01
Workflow shashank

yadavshashankr:


app/src/main/java/com/portfolio/tasky/ToolbarHandlerImpl.kt
```diff
@@ -0,0 +1,23 @@
+package com.portfolio.tasky
+
+import android.view.animation.TranslateAnimation
+import androidx.appcompat.widget.Toolbar
+import com.portfolio.tasky.globals.Constants
+
+
+class ToolbarHandlerImpl : ToolbarHandler {
```
MKiperszmid, line 8: I like that you did this

But I got a question

In which part of the Tasky app will you use a Toolbar?

app/src/main/java/com/portfolio/tasky/ToolbarHandlerImpl.kt
```diff
@@ -0,0 +1,23 @@
...
+    private lateinit var mToolbar: Toolbar
+    private lateinit var translateAnimation: TranslateAnimation
+
+    override fun setToolBarText(toolbar: Toolbar, text: String) {
+        mToolbar = toolbar
```
MKiperszmid, line 13: You are saving a variable "mToolbar" but then are never using it
"mToolbar.title" could just be "toolbar.title" (Toolbar from parameter)

So you could literally remove the variable mToolbar ;)

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -3,20 +3,30 @@ package com.portfolio.tasky.entry
...
-import com.portfolio.tasky.R
+import androidx.appcompat.widget.Toolbar
+import androidx.databinding.DataBindingUtil
+import com.portfolio.tasky.*
+import com.portfolio.tasky.databinding.ActivityEntryBinding
 import com.portfolio.tasky.entry.fragments.LoginFragment
 
-class EntryActivity : AppCompatActivity(), FragmentInflater by FragmentInflaterImpl() {
+class EntryActivity : AppCompatActivity(), FragmentInflater by FragmentInflaterImpl(), ToolbarHandler by ToolbarHandlerImpl() {
+    private var viewBinding: ActivityEntryBinding? = null
```
MKiperszmid, line 13: viewBinding could be lateinit

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @SuppressLint("CustomViewStyleable")
+    private fun setupView(attrs: AttributeSet) {
+        val inflater = context.getSystemService(Context.LAYOUT_INFLATER_SERVICE) as LayoutInflater
+        viewBinding = LayoutInputFieldBinding.inflate(inflater, this, false)
+        addView(viewBinding.root)
+
+        editText = viewBinding.etInput
+        val ta: TypedArray = context.obtainStyledAttributes(attrs, R.styleable.EditTextInputField)
```
MKiperszmid, line 55: Make variable names more readable

``` val typedArray ``` is way easier to read than ``` val ta ```

**Review by MKiperszmid — COMMENTED** [review]
Some minor comments

File: app/src/main/java/com/portfolio/tasky/ToolbarHandlerImpl.kt (hunk 1 above)
Comment (yadavshashankr) [reply app/src/main/java/com/portfolio/tasky/ToolbarHandlerImpl.kt:8]: In Login, Registration, Task Details, Event Detail and Reminder Detail screens. 

Since there are dates to be shown in the Tasks, Events and Reminder screens, I'll be able to utilise the current implementation for setting the text and animation.

File: app/src/main/java/com/portfolio/tasky/ToolbarHandlerImpl.kt [hunk 5]
```diff
@@ -0,0 +1,23 @@
+package com.portfolio.tasky
+
+import android.view.animation.TranslateAnimation
+import androidx.appcompat.widget.Toolbar
+import com.portfolio.tasky.globals.Constants
+
+
+class ToolbarHandlerImpl : ToolbarHandler {
+    private lateinit var mToolbar: Toolbar
```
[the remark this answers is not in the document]
yadavshashankr: Okay, thanks for the valuable info !

Done

app/src/main/java/com/portfolio/tasky/ToolbarHandlerImpl.kt
yadavshashankr: True! 
Done

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
yadavshashankr: Yes!

Done

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/globals/Constants.kt
```diff
@@ -0,0 +1,12 @@
+package com.portfolio.tasky.globals
+
+class Constants {
```
philipplackner, line 3: I'd personally use more grouped classes for constants and not one global constants class. In the end you'll have different features and each will have its own set of constants. You already kinda did that by having different objects in this class, but to be able to have a clear separation of concerns, it doesn't work to have a global place for this

PR #3, opened 2023-03-03
Workflow shashank

yadavshashankr:


yadavshashankr:
https://user-images.githubusercontent.com/81215694/222853356-34d94cfc-656a-4e9a-9331-8a95ae8396f8.mp4

app/src/main/AndroidManifest.xml
```diff
@@ -15,6 +15,7 @@
         tools:targetApi="31">
         <activity
             android:name=".entry.EntryActivity"
+            android:windowSoftInputMode="stateAlwaysHidden"
```
MKiperszmid, line 18: Why did you use this one?
Wouldn't this mean that you'll have to manually open the keyboard?

app/src/main/java/com/portfolio/tasky/TaskyValidationWatcherImpl.kt
```diff
@@ -0,0 +1,64 @@
...
+    override fun beforeTextChanged(p0: CharSequence, p1: Int, p2: Int, p3: Int) {/*TODO("Not yet implemented")*/}
+
+    override fun onTextChanged(p0: CharSequence, p1: Int, p2: Int, p3: Int) {
+        validateText(p0, taskAppCompatEditText.id, taskAppCompatEditText)
+    }
+
+    override fun afterTextChanged(p0: Editable?) {/*TODO("Not yet implemented")*/}
+
+    private fun validateText(var1: CharSequence, viewID: Int, taskAppCompatEditText: TaskyAppCompatEditText) {
+        if(viewID == R.id.et_email){
```
MKiperszmid, line 39: You are hardcoding the ids on the validator

What would happen if for example

You add this validator on Registration? Where they require a name
Or when you add a new attendee?

This means you'll have to always remember to add the id in this method

What I recommend, is having specific usecases
Ex: ValidateEmailUseCase, ValidatePasswordUseCase
And call it from a viewmodel
This means, everytime you have to use an email, you can just pass the EmailValidator, and that's it
No need to use IDs or anything

app/src/main/java/com/portfolio/tasky/TaskyValidationWatcherImpl.kt
```diff
@@ -0,0 +1,64 @@
...
+        }
+        buttonActivator()
+    }
+
+    private fun buttonActivator() {
+        val isValid =
+       /*Login*/      isEntryValid == LOGIN && isEmailValid && isPasswordValid
+         /*OR*/          ||
+    /*Registration*/  isEntryValid == REGISTRATION && isEmailValid && isPasswordValid && isNameValid
+        fieldsValidator?.fieldsValidated(isValid)
```
MKiperszmid, line 62: If you end up using UseCase and a ViewModel, you could have a state on the viewmodel that's "isValidInformation", and set it to true or false based on the UseCases

And then the button would just observe that state

app/src/main/java/com/portfolio/tasky/FieldsValidator.kt
```diff
@@ -0,0 +1,5 @@
+package com.portfolio.tasky
+
+interface FieldsValidator {
+    fun fieldsValidated(valid : Boolean)
```
MKiperszmid, line 4: If you end up using UseCases (Explained on other comments)

This wouldn't be necessary. Since the usecases will return a Boolean themselves

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -33,24 +35,85 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl() {
...
         }
     }
+    private fun setFieldValidations(loginFragment: LoginFragment?) {
+        clearValidatorParams()
+
+        val etEmail = viewBinding.etEmail.subLayout.etInput
+        etEmail.addTextChangedListener(loginFragment?.let {
+            TaskyValidationWatcherImpl(viewBinding.etEmail, LOGIN,
+                it
+            )
```
MKiperszmid, line 45: Here you'd just call the viewmodel.onEmailChange or something

And the viewmodel would just call the emailvalidator

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -33,24 +35,85 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl() {
...
+
+    override fun fieldsValidated(valid: Boolean) {
+        viewBinding.btnLogin.isEnabled = valid
+        valid.let {
+            if (it) {
+                viewBinding.etEmail.clearFocus()
+                viewBinding.etPassword.clearFocus()
+                viewBinding.btnLogin.requestFocus()
+            }
+        }
```
MKiperszmid, line 82: There's an issue with this method. It's something I noticed on the video you added (Great you added video, love it)

When the email is valid, and then the password textfield gets verified, it's removing the focus.

No need to remove the focus in this case, since the user might be writing a longer password

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -33,24 +35,85 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl() {
...
-            val loginFragment = LoginFragment()
-            loginFragment.apply {
-                enterTransition = Slide(Gravity.BOTTOM)
-                exitTransition = Fade(Fade.MODE_OUT)
-            }
-            return loginFragment
+                if (loginFragment == null){
+                    loginFragment = LoginFragment()
+                }
+                loginFragment?.apply {
```
MKiperszmid, line 112: LoginFragment won't be null here, since you asked if it's null before, and then created it

app/src/main/java/com/portfolio/tasky/globals/DataUtils.kt
```diff
@@ -0,0 +1,30 @@
...
+            }
+
+            fun isEmailValid(emailID : CharSequence?) : Boolean{
+                return Patterns.EMAIL_ADDRESS.matcher(emailID as CharSequence).matches()
+            }
+            fun isPasswordValid(password : CharSequence) : Boolean{
+                return Regex("^(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z]).{9,}\$").matches(password)
+            }
+            fun isNameValid(name : CharSequence) : Boolean{
+                return Regex("^(?=.*[a-z]).{3,}\$").matches(name)
```
MKiperszmid, line 26: These can all be usecases
Will make more sense, since each would be in their own class (Check the video about SOLID in the video portal of the mentorship, the S specifically)

Also, using Patterns.EMAIL_ADDRESS will make it so the method isn't Unit Testable since it's an Android-specific dependency. You can check the import in like 3 "import android.util.Patterns"
I recommend:

1. You create an interface
2. You write an implementation of that, which uses Patterns.EMAIL_ADDRESS
3. You pass the abstraction to the constructor of your class needing it(The ValidateEmailUseCase), or if you want to keep this class, then you can pass it on the constructor or the parameter

An example:

```
interface EmailPatternValidator {
    fun isValidEmailPattern(email: String): Boolean
}
```

```
class EmailPatternValidatorImpl: EmailPatternValidator {
    override fun isValidEmailPattern(email: String): Boolean {
        return Patterns.EMAIL_ADDRESS.matcher(email).matches()
    }
}
```

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
```diff
@@ -1,62 +1,117 @@
...
         error = typedArray.getBoolean(R.styleable.EditTextInputField_error, false)
         drawableLast = typedArray.getDrawable(R.styleable.EditTextInputField_drawableLast)
         inputType = typedArray.getInt(R.styleable.EditTextInputField_inputType, InputType.TYPE_CLASS_TEXT)
         hint = typedArray.getString(R.styleable.EditTextInputField_hint)
+        isPassword = typedArray.getBoolean(R.styleable.EditTextInputField_isPassword, false)
         typedArray.recycle()
     }
+
+    override fun onTouch(v: View, event: MotionEvent): Boolean {
+        when (event.action) {
```
MKiperszmid, line 88: If you are just going to use one event action, it can just be an if statement

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
```diff
@@ -1,62 +1,117 @@
...
+            MotionEvent.ACTION_UP -> {
+                showHidePassword(event)
+            }
+        }
+        return false
+    }
+
+    private fun showHidePassword(event: MotionEvent) {
+        val textLocation = IntArray(2)
+        if (event.rawX >= textLocation[0] + subLayout.etInput.width - subLayout.etInput.totalPaddingRight) {
```
MKiperszmid, line 98: What does this if statement do?

I recommend creating a variable that describes what it's doing, and then checking that variable

Example:
```
val isValidPositionForInput = event.rawX >= textLocation[0] + subLayout.etInput.width - subLayout.etInput.totalPaddingRight
```

This is just an example, you'd name it based on what it does

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
```diff
@@ -1,62 +1,117 @@
...
+
+            if(subLayout.etInput.compoundDrawablesRelative[2] != null && !subLayout.valid!!) {
```
MKiperszmid, line 100: Same here

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
```diff
@@ -1,62 +1,117 @@
...
+
     init {
         setupView(attrs)
-        isError()
+        setError(false)
         setDrawableLast()
         setInputType()
         setHint()
+        setValid(false)
+        isPassword()
```
MKiperszmid, line 40: This is just an idea
Maybe make a different EditText, that inherits from this one, but make it only for password

Something like

TaskyAppCompatPasswordEditText

So you have 2 edit texts with different logic

To be 100% honest, I haven't done many custom views on XML (I did couple years ago) so I might be a bit rusty with it
But should work

MKiperszmid (approved):
Couple of comments

app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
yadavshashankr: Yes. I wanted to hide the keyboard when the new fragment loads. This required fields to not have focus and hide the keyboard automatically..

app/src/main/java/com/portfolio/tasky/FieldsValidator.kt
yadavshashankr: Okay.
Done.

app/src/main/java/com/portfolio/tasky/TaskyValidationWatcherImpl.kt
yadavshashankr: True..

Done

app/src/main/java/com/portfolio/tasky/TaskyValidationWatcherImpl.kt
yadavshashankr: True..

Instead of a state, I have created a mutable object and listening to both the values being true in login..in case of registration all the three values to be true. 

Also if you could recommend me some video explaining "State" for beginners.

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
yadavshashankr: Correct. Implemented as per this.

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
yadavshashankr: Correct. Rectified it!

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
yadavshashankr: Yes, I was trying to do some RnD and forgot to re-fix the code.

Done

app/src/main/java/com/portfolio/tasky/globals/DataUtils.kt
yadavshashankr: Thank you for explaining in this detail. Your explanation helped me implement these use-cases.

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
yadavshashankr: We can use another AppCompatEdittext for Passwords but that would just require to handle few more files. In the current implementation, the custom field handles both inputTypes, TEXT & PASSWORD.

yadavshashankr:
https://user-images.githubusercontent.com/81215694/222923517-de0ecee7-f80e-48f9-ad4f-a3747c81ed4a.mp4

app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
MKiperszmid: Gotcha

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
yadavshashankr: It actually makes the end area of AppCompatEditText clickable.

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -1,56 +1,137 @@
...
+        super.onResume()
+        setFieldValidations(this)
+    }
+    override fun onDestroyView() {
+        super.onDestroyView()
+        removeFieldValidations(this)
+    }
+
     companion object {
+        private var loginFragment : LoginFragment? = null
```
MKiperszmid, line 127: Make it lateinit, so you avoid nullability since you are always creating the instance

So:
``` 
private lateinit var loginFragment: LoginFragment
```

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -1,56 +1,137 @@
...
+            val passwordField = viewBinding.etPassword
+            it?.let { it1 -> passwordField.setValid(it1)
+                passwordField.setError(passwordField.subLayout.etInput.text?.isNotEmpty() as Boolean && !it1)}
+        }
+        viewModel.areFieldsValid.observe(viewLifecycleOwner){
+            it.let {  viewBinding.btnLogin.isEnabled = it == true }
+        }
+    }
+
+    private fun setFieldValidations(loginFragment: LoginFragment?) {
```
MKiperszmid, line 61: Why is LoginFragment nullable?

I made another comment, maybe that fixes the nullability

app/src/main/java/com/portfolio/tasky/usecases/EmailPatternValidatorImpl.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import android.util.Patterns
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class EmailPatternValidatorImpl @Inject constructor() : EmailPatternValidator {
+    override fun isValidEmailPattern(email: CharSequence?): Boolean {
+        return Patterns.EMAIL_ADDRESS.matcher(email as CharSequence).matches()
+    }
```
MKiperszmid, line 11: This implementation should be in the "data" portion of the feature

Check the video about "Proper Package Structure" (or however it's called) to know more about how to structure the app

app/src/main/java/com/portfolio/tasky/usecases/EmailPatternValidator.kt
```diff
@@ -0,0 +1,5 @@
+package com.portfolio.tasky.usecases
+
+interface EmailPatternValidator {
+    fun isValidEmailPattern(email : CharSequence?) : Boolean?
```
MKiperszmid, line 4: No need to add that many nullables

You could just do
```
fun isValidEmailPattern(email : String) : Boolean
```

app/src/main/java/com/portfolio/tasky/usecases/NamePatternValidator.kt
```diff
@@ -0,0 +1,5 @@
+package com.portfolio.tasky.usecases
+
+interface NamePatternValidator {
+    fun isValidName(name: CharSequence?) : Boolean?
```
MKiperszmid, line 4: Same as what I mentioned in the EmailPatternValidator comment

app/src/main/java/com/portfolio/tasky/usecases/NamePatternValidatorImpl.kt
```diff
@@ -0,0 +1,9 @@
+package com.portfolio.tasky.usecases
+
+import javax.inject.Inject
+
+class NamePatternValidatorImpl @Inject constructor() : NamePatternValidator {
+    override fun isValidName(name: CharSequence?): Boolean {
+        return Regex("^(?=.*[a-z]).{3,}\$").matches(name as CharSequence)
```
MKiperszmid, line 7: In this case, since you are going to use your own regex, you don't need to have a NamePatternValidator and a NamePatternValidatorImpl
Since it's fully Kotlin code, nothing related to Android

Also, no need to use regex

You could simply check for stuff like "isNotEmpty" and "Length > 5" or similar stuff

app/src/main/java/com/portfolio/tasky/usecases/PasswordPatternValidatorImpl.kt
```diff
@@ -0,0 +1,9 @@
+package com.portfolio.tasky.usecases
+
+import javax.inject.Inject
+
+class PasswordPatternValidatorImpl @Inject constructor(): PasswordPatternValidator {
```
MKiperszmid, line 5: Same as "NamePatternValidatorImpl" comment

+ here you would check for:
At least 1 lower case letter
At least 1 upper case letter
At least 1 number
Minimum length 8 (or 9, can't really remember)

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.lifecycle.ViewModel
+import com.portfolio.tasky.usecases.EmailPatternValidatorImpl
+import com.portfolio.tasky.usecases.PasswordPatternValidatorImpl
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+
+@HiltViewModel
+class LoginViewModel @Inject constructor(
+    private val validatorImpl: EmailPatternValidatorImpl,
+    private val passwordPatternValidatorImpl: PasswordPatternValidatorImpl
```
MKiperszmid, line 14: Here you should send the abstractions as parameter, but inject the implementation in the ValidatorModule

This is useful because when testing the viewmodel, you could pass "FakeEmailValidator" or something

MKiperszmid: Also, the names of the email validator isn't that descriptive
"validatorImpl" doesn't provide information

I'd go with something like
"emailValidator"

Alslo, try not to add "Impl" to the variable names

So for the password one, i'd fo with
"passwordValidator"

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
```diff
@@ -0,0 +1,35 @@
...
+): ViewModel() {
+    private val mutableEmailChange = MutableLiveData<Boolean?>()
+    val emailChange : LiveData<Boolean?> = mutableEmailChange
+
+    private val mutablePasswordChange = MutableLiveData<Boolean?>()
+    val passwordChange : LiveData<Boolean?> = mutablePasswordChange
+
+    private val mutableFieldsValid = MutableLiveData<Boolean?>()
+    val areFieldsValid : LiveData<Boolean?> = mutableFieldsValid
```
MKiperszmid, line 23: Try to avoid nulls here
You could start with false as default, meaning they are invalid at the start

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    fun emailChange(email : CharSequence?){
+        mutableEmailChange.value = validatorImpl.isValidEmailPattern(email)
+    }
+    fun passwordChange(password : CharSequence?){
+        mutablePasswordChange.value = passwordPatternValidatorImpl.isPasswordPatternValid(password)
+    }
+    fun areFieldsValid(){
+        mutableFieldsValid.value =  (mutableEmailChange.value != null && mutablePasswordChange.value == true) && (mutablePasswordChange.value != null && mutablePasswordChange.value == true)
```
MKiperszmid, line 32: Since you removed nullability (explained on my previous comment), this code will be way more readable

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
MKiperszmid: Sounds good. Let's leave it like it is then

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -1,56 +1,137 @@
...
+        val etEmail = viewBinding.etEmail.subLayout.etInput
+        val etPassword = viewBinding.etPassword.subLayout.etInput
+
+        etEmail.addTextChangedListener(loginFragment?.let {
+            TaskyValidationWatcherImpl(viewBinding.etEmail,
+                it
+            )
```
MKiperszmid, line 68: Here I prefer calling the viewmodel, and have the viewmodel handle all the validation calls

UIs (Fragments, Activities, etc) should be as dumb as possible. They should just display the state, and on state change, they should call the viewmodel

So maybe create a 
"onEmailChanged" function on the viewmodel. And there, the viewmodel will call the email validator use case

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -1,56 +1,137 @@
...
+        })
+        etPassword.addTextChangedListener(loginFragment?.let {
+            TaskyValidationWatcherImpl(viewBinding.etPassword,
+                it
+            )
```
MKiperszmid, line 73: Same as what I mentioned in the email comment

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -1,56 +1,137 @@
...
+        if (taskyAppcompatEditText.id == R.id.et_email){
+            viewModel.emailChange(editable.toString() as CharSequence)
+        }
+
+
+        if(taskyAppcompatEditText.id == R.id.et_password){
+            viewModel.passwordChange(editable.toString() as CharSequence)
+        }
+            viewModel.areFieldsValid()
+    }
```
MKiperszmid, line 87: If you do what I tell you in the email comment, then this method and the whole interface can be deleted ;)

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -1,56 +1,137 @@
...
+        setObservers()
         viewBinding.btnLogin.setOnClickListener {
             openRegistrationFragment()
         }
     }
 
+    private fun setObservers() {
+        viewModel.emailChange.observe(viewLifecycleOwner){
+            val emailField = viewBinding.etEmail
+            it?.let { it1 -> emailField.setValid(it1)
```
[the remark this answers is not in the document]
yadavshashankr: Okay 

Done

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -1,56 +1,137 @@
...
+                emailField.setError(emailField.subLayout.etInput.text?.isNotEmpty() as Boolean && !it1)}
```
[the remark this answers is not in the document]
yadavshashankr: Had took it for something else..

Done

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
yadavshashankr: Some experimentation and then forgot to clean it up

Done

app/src/main/java/com/portfolio/tasky/usecases/EmailPatternValidatorImpl.kt
yadavshashankr: Sure I ill look at the structure video again but if it's ok then we can skip this one for now. I'll rectify the package structure in the next PR since it will require only refactoring of file locations.

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
yadavshashankr: True

Rectified it!

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
yadavshashankr: I am using a Custom text watcher to listen to text changes, since I have two fields in my layout, I have to check for their ID's in order to activate the correct viewModel.onEmailChange or viewModel.onPasswordChange..

I am not able to comprehend this..please if possible then direct me to some code or video

yadavshashankr: "onEmailChanged" function on the viewmodel. And there, the viewmodel will call the email validator use case


I think this is how it has been implemented..

app/src/main/java/com/portfolio/tasky/usecases/PasswordPatternValidatorImpl.kt
yadavshashankr: Yes , this is what the regex has been configured for. 1 number, 1 upper-case, 1 lowe-case and minimum length 9.

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/di/modules/ValidatorsModule.kt
```diff
@@ -0,0 +1,24 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+class ValidatorsModule {
+        @Provides
+        @Singleton
+        fun provideEmailPatternValidationImpl(): EmailPatternValidator {
```
MKiperszmid, line 15: Try not to add "Impl" to method names either
So in this case it would be 

"provideEmailPatternValidation"

app/src/main/java/com/portfolio/tasky/di/modules/ValidatorsModule.kt
```diff
@@ -0,0 +1,24 @@
...
+                return EmailPatternValidatorImpl()
+        }
+
+        @Provides
+        @Singleton
+        fun providePasswordPatternValidationImpl(): PasswordPatternValidator {
```
MKiperszmid, line 21: Same here

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
```diff
@@ -1,55 +1,169 @@
...
+                it
+            )
+        })
+    }
+
+
+
+    override fun onTextChanged(editable: Editable, taskyAppcompatEditText: TaskyAppCompatEditText) {
+        if (taskyAppcompatEditText.id == R.id.et_email){
+            viewModel.emailChange(editable.toString() as CharSequence)
```
MKiperszmid, line 92: Why are you converting to CharSequence?

app/src/main/java/com/portfolio/tasky/usecases/EmailPatternValidator.kt
```diff
@@ -0,0 +1,5 @@
+package com.portfolio.tasky.usecases
+
+interface EmailPatternValidator {
+    fun isValidEmailPattern(email : CharSequence) : Boolean
```
MKiperszmid, line 4: Why not leave it as String instead of CharSequence?

app/src/main/java/com/portfolio/tasky/usecases/PasswordPatternValidator.kt
```diff
@@ -0,0 +1,5 @@
+package com.portfolio.tasky.usecases
+
+interface PasswordPatternValidator {
```
MKiperszmid, line 3: This, as well as the FullNameValidator, doesn't need to be an abstraction

Since they don't require any Android-specific dependencies

The only one that did, is the email one
That's why that one needs to have an abstraction

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
```diff
@@ -0,0 +1,38 @@
...
+class LoginViewModel @Inject constructor(
+    private val emailValidator: EmailPatternValidator,
+    private val passwordPatternValidator: PasswordPatternValidator
+): ViewModel() {
+
+    private val mutableEmailChange = MutableLiveData(false)
+    val emailChange : LiveData<Boolean> = mutableEmailChange
+
+    private val mutablePasswordChange = MutableLiveData(false)
+    val passwordChange : LiveData<Boolean?> = mutablePasswordChange
```
MKiperszmid, line 21: You can change Boolean? for Boolean

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    private val mutableFieldsValid = MutableLiveData(false)
+    val areFieldsValid : LiveData<Boolean?> = mutableFieldsValid
```
MKiperszmid, line 24: You can change Boolean? for Boolean

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
```diff
@@ -0,0 +1,38 @@
...
+
+
+
+    fun emailChange(email : CharSequence){
```
MKiperszmid, line 28: I recommend leaving it as String

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/viewModels/LoginViewModel.kt
yadavshashankr: Forgot this one as well ::face_palm::

Done

app/src/main/java/com/portfolio/tasky/usecases/EmailPatternValidator.kt
yadavshashankr: Sry, didn't gave much thought to this.

Done

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/di/modules/ValidatorsModule.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/di/modules/ValidatorsModule.kt
yadavshashankr: Okay

Done

PR #4, opened 2023-03-05
Workflow shashank

yadavshashankr:


app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -43,17 +47,17 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl(), Te
     }
 
     private fun setObservers() {
-        viewModel.emailChange.observe(viewLifecycleOwner){
+        viewModel.emailChange.observe(viewLifecycleOwner) {
```
philipplackner, line 50: Instead of naming it `emailChange` I'd prefer something like just `email` since a LiveData doesn't reflect an email change really, but rather just holds the value of that email for which you can then subscribe to changes here.

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -43,17 +47,17 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl(), Te
...
             val emailField = viewBinding.etEmail
             it?.let { isValid -> emailField.setValid(isValid)
                 emailField.setError(emailField.subLayout.etInput.text?.isNotEmpty() == true && !isValid)}
         }
-        viewModel.passwordChange.observe(viewLifecycleOwner){
+        viewModel.passwordChange.observe(viewLifecycleOwner) {
```
philipplackner, line 55: Same here

app/src/main/java/com/portfolio/tasky/entry/usecases/PasswordPatternValidation.kt
```diff
@@ -1,8 +1,8 @@
-package com.portfolio.tasky.usecases
+package com.portfolio.tasky.entry.usecases
 
 import javax.inject.Inject
 
-class PasswordPatternValidation @Inject constructor(){
+class PasswordPatternValidation @Inject constructor() {
     fun isPasswordPatternValid(password: CharSequence): Boolean {
         return Regex("^(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z]).{9,}\$").matches(password)
     }
```
philipplackner, line 8: I would personally avoid regex if not really necessary. Nobody will be able to quickly tell you what this matches for. A function like this would be much more readable:
```
val hasValidLength = password.length in (MIN_CHARACTERS_FOR_PASSWORD..MAX_CHARACTERS_FOR_PASSWORD)
val hasUppercase = password.any { it.isUppercase() }
val hasLowercase = password.any { it.isLowercase() }
val hasDigits = password.any { it.isDigit() }

return hasValidLength && hasUppercase && hasLowercase && hasDigits
```

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
```diff
@@ -25,13 +25,13 @@ class LoginViewModel @Inject constructor(
...
-    fun emailChange(email : String){
+    fun emailChange(email : String) {
         mutableEmailChange.value = emailValidator.isValidEmailPattern(email)
     }
-    fun passwordChange(password : String){
+    fun passwordChange(password : String) {
         mutablePasswordChange.value = passwordPatternValidation.isPasswordPatternValid(password)
     }
-    fun areFieldsValid(){
+    fun areFieldsValid() {
```
philipplackner, line 34: This function sounds like it returns a boolean which it doesn't. The fact that it updates a LiveData is considered a side-effect, so something you don't expect happens when reading the function name. 2 ways how you could improve that:
1. Rename the function to something like `validateFields()` or `checkIfFieldsValid()` which would make this more clear
2. You return `mutablePasswordChange.value == true && mutablePasswordChange.value == true` in this function and call it from somewhere else when you want to update the LiveData 

I personally prefer option 1.

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
```diff
@@ -25,13 +25,13 @@ class LoginViewModel @Inject constructor(
 
 
 
-    fun emailChange(email : String){
+    fun emailChange(email : String) {
```
philipplackner, line 28: I'd always try to name functions based on what they do. `emailChange` isn't really something you do. Might sound nitpicky here, but in larger codebases these things actually matter quite a lot, since new devs need to quickly understand parts of a codebase.
In this case, something like `onEmailChange()` or `submitEmailChange()` or so would make the purpose more clear 👍

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
```diff
@@ -25,13 +25,13 @@ class LoginViewModel @Inject constructor(
...
         mutableEmailChange.value = emailValidator.isValidEmailPattern(email)
     }
-    fun passwordChange(password : String){
+    fun passwordChange(password : String) {
```
philipplackner, line 31: Same feedback here

app/src/main/java/com/portfolio/tasky/entry/viewModels/RegisterViewModel.kt
```diff
@@ -30,16 +30,16 @@ class RegisterViewModel  @Inject constructor(
...
+    fun emailChange(email : String) {
         mutableEmailChange.value = emailValidator.isValidEmailPattern(email)
     }
-    fun passwordChange(password : String){
+    fun passwordChange(password : String) {
         mutablePasswordChange.value = passwordPatternValidation.isPasswordPatternValid(password)
     }
-    fun nameChange(name : String){
-        mutableNameChange.value =nameValidation.isValidName(name)
+    fun nameChange(name : String) {
```
philipplackner, line 39: Same feedback here

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
```diff
@@ -53,17 +53,18 @@ class TaskyAppCompatEditText(context: Context, attrs: AttributeSet) : FrameLayou
...
+    fun setError(error : Boolean) {
         subLayout.error = error
     }
 
-    fun setValid(valid: Boolean){
+    fun setValid(valid: Boolean) {
         subLayout.valid = valid
     }
-    private fun isPassword(){
+    private fun isPassword() {
```
philipplackner, line 63: Same feedback as before here - as soon as a function starts with `is` or `are` it's formulated as a question which you expect it to answer (by returning a boolean in this case). However this function doesn't do that so from the outside it's kinda cryptic what it does.
And to be honest, I don't really understand what it even does, since you assign something that sounds like a boolean (`isPassword`) to a text (`subLayout.password`). Could you clarify that? 👍

philipplackner (approved):

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt

app/src/main/java/com/portfolio/tasky/entry/usecases/PasswordPatternValidation.kt
yadavshashankr: True

Thanks for providing a better readable fun! Done

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
yadavshashankr: Okay

Done

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
yadavshashankr: Okay

Done

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/entry/viewModels/RegisterViewModel.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/views/TaskyAppCompatEditText.kt
yadavshashankr: True, the method name doesn't describe its purpose properly. 

The custom edittext's inputType is switched here, by providing 'isPassword' value to a data binding object of type boolean, when clicked on the eye icon. 

This data binding object switches the inputType of inner AppCompatEdittext inputType to 'TEXT_CLASS_TEXT / TEXT_CLASS_VARIATION_PASSWORD'

philipplackner (approved):

PR #5, opened 2023-03-05
Network state and online/offline mode identifier

yadavshashankr:
1. Implementing Internet / Network state.
2. Global Offline/Online mode and identifier.


yadavshashankr:
https://user-images.githubusercontent.com/81215694/222981286-0579af96-26fb-45a9-ba4f-6af62a9130bc.mp4

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -33,6 +41,33 @@ class EntryActivity : AppCompatActivity(), FragmentInflater by FragmentInflaterI
...
-        setToolBarText(viewBinding?.toolbar as Toolbar, text)
+        setToolBarText(viewBinding.toolbar, text)
+    }
+
+    override fun internetAvailable() {
+
+        runOnUiThread {
+            viewBinding.layoutOnlineMode.tvOnlineMode.let { onConnected(it, true) }
+            viewBinding.layoutOnlineMode.drawableFirst = ContextCompat.getDrawable(this, R.drawable.ic_online)
+        }
```
MKiperszmid, line 52: Isn't this already running on ui thread?

yadavshashankr:
https://user-images.githubusercontent.com/81215694/222987595-2c9ac966-2121-44f3-9879-ed9936d5e006.mp4


Listening to internet connectivity without changing network. Event time in the video at 00:16 and 00:37

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
```diff
@@ -0,0 +1,62 @@
...
+import android.content.Intent
+import android.net.ConnectivityManager
+import android.net.ConnectivityManager.NetworkCallback
+import android.net.Network
+import android.net.NetworkCapabilities
+import android.net.NetworkRequest
+
+
+
+
```
MKiperszmid, line 14: Extra spacing

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
```diff
@@ -0,0 +1,62 @@
...
+
+class NetworkChangeReceiver constructor(internetStatus: InternetStatus) : BroadcastReceiver() {
```
MKiperszmid, line 16: Remember, if a usecase uses android-specific dependencies, then is should be an interface, where the implementation takes care of the android part. But the interface doesn't have Android logic

Check: https://www.youtube.com/watch?v=TzV0oCRDNfM

app/src/main/java/com/portfolio/tasky/usecases/OfflineModeHandler.kt
```diff
@@ -0,0 +1,7 @@
+package com.portfolio.tasky.usecases
+
+import androidx.appcompat.widget.AppCompatTextView
+
+interface OfflineModeHandler {
```
MKiperszmid, line 5: The usecase would be the "Network Change Receiver", this one just modifies the UI. So this can be implemented directly on the UI

MKiperszmid (approved):
Couple of comments

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
yadavshashankr: Here I have used NetworkChangeReceiver as my impl class, I used InternetStatus interface as abstraction. When a network event generates in NetworkChangeReceiver my EntryActivity class receives it through abstraction/interface's overridden methods, internetAvailable(), internetNotAvailable() ..I think for testing I can use some FakeInternetStatus interface to test it out..

I am not sure what I am missing 😕

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
yadavshashankr: Yes, but it was crashing. From what I understand onAvailable and onNetworkCapabilitiesChanges from NetworkCallback sometimes run on Worker thread due to race conditions.  

Anyways with new implementation this code has been removed

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
yadavshashankr: Now I get it. Earlier I had used interface but couldn't get it working, with new implementation I have used LiveData which has a type of an interface called NetworkStatus. I hope this is allowed..

app/src/main/java/com/portfolio/tasky/usecases/OfflineModeHandler.kt
yadavshashankr: Since all this impl does was change UI so have implemented in the EntryActivity itself.

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -2,28 +2,75 @@ package com.portfolio.tasky.entry
...
 
+    private fun setObservers() {
+        networkChangeReceiver = NetworkChangeReceiver(this)
+        networkChangeReceiver.observe(this){
+            when(it){
+                NetworkStatus.Available -> networkAvailability(true)
+                NetworkStatus.Unavailable -> networkAvailability(false)
+            }
+        }
+    }
```
philipplackner, line 40: I'd keep the activities and fragments as dumb as possible. The only thing they should really do is observe states from a ViewModel and update the Views based on that (and things like navigation, permission handling or similar). 

So in this case, I'd let the ViewModel observe the network state and just forward the state to the UI which can then decide what it does with it.

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -2,28 +2,75 @@ package com.portfolio.tasky.entry
...
+
+    private fun networkAvailability(isAvailable : Boolean){
```
philipplackner, line 42: Similar comment as in a previous PR: I'd keep function names as clear as possible, so here I'd rather use `onNetworkAvailabilityChanged()` or similar

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -2,28 +2,75 @@ package com.portfolio.tasky.entry
...
+        onConnected(viewBinding.layoutOnlineMode.tvOnlineMode, isAvailable)
+        if(isAvailable){
+            viewBinding.layoutOnlineMode.drawableFirst = ContextCompat.getDrawable(this, R.drawable.ic_online)
+        }else{
+            viewBinding.layoutOnlineMode.drawableFirst = ContextCompat.getDrawable(this, R.drawable.ic_offline)
+        }
+    }
+
+     private fun onConnected(offlineTextField: AppCompatTextView, connected: Boolean) {
```
philipplackner, line 51: Why do you need to pass the `offlineTextField` here. This function is part of your activity and should have access to that textfield right?

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -2,28 +2,75 @@ package com.portfolio.tasky.entry
...
+        if (connected){
+            offlineTextField.text = Constants.Companion.AppConnectivityMode.ONLINE
+        }else{
+            offlineTextField.text = Constants.Companion.AppConnectivityMode.OFFLINE
```
philipplackner, line 55: Are these constants hardcoded strings? If so, I'd prefer using string resources, since they allow you to later localize/translate your app.

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -2,28 +2,75 @@ package com.portfolio.tasky.entry
...
+    private fun animate(offlineTextField: AppCompatTextView, connected: Boolean) {
+        offlineTextField.visibility = View.VISIBLE
+        val endToStartAnimation = TranslateAnimation(500f, 0f, 0f, 0f)
+        endToStartAnimation.duration = Constants.Companion.AnimationProperties.DURATION
+        offlineTextField.startAnimation(endToStartAnimation)
+
+        Handler(Looper.getMainLooper()).postDelayed({
+            offlineTextField.text = ""
+            if (connected){
+                offlineTextField.visibility =  View.GONE
```
philipplackner, line 69: Instead of this, you can also just do `offlineTextField.isVisible = false` (false for GONE, true for VISIBLE). It's usually easier to work with that since you can directly assign a boolean to it.

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    override fun onActive() {
+        isOnline()
+    }
+
+    override fun onInactive() {
+        connectivityManager.unregisterNetworkCallback(networkCallback)
+    }
+
+    private fun isOnline() {
```
philipplackner, line 25: Same feedback as in previous PR: This function doesn't tell me what it does. I'd expect that it returns a boolean which it doesn't, so it's full of unknown side-effects from the outside.

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
```diff
@@ -0,0 +1,68 @@
...
+
+
+class NetworkChangeReceiver(var context: Context) : LiveData<NetworkStatus>() {
+
+    val validNetworkConnections : MutableSet<Network> = HashSet()
+    private var connectivityManager: ConnectivityManager =
+        context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+
+    override fun onActive() {
+        isOnline()
```
philipplackner, line 18: Since your code from `isOnline()` is only used here, you could also directly put it in `onActive()`

app/src/main/java/com/portfolio/tasky/usecases/NetworkStatus.kt
```diff
@@ -0,0 +1,6 @@
+package com.portfolio.tasky.usecases
+
+sealed class NetworkStatus{
+    object Available : NetworkStatus()
+    object Unavailable : NetworkStatus()
+}
```
philipplackner, line 6: If you only need these 2 statuses, you can also just use a boolean `isAvailable` for the sake of simplicity. Also, this could be an `enum class` since your different statuses don't need different constructors (which would be a reason to use a sealed class)

philipplackner (approved):

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
yadavshashankr: I have implemented EntryViewModel and currently observing network change from there. Only updating views in Activity class like setting text, animating and changing visibility of views.

Earlier also the NetworkChangeReceiver was extended to LiveData<NetworkStatus>, and I was observing it by initialising it in the activity..I thought we can avoid creating an extra ViewModel since the NetworkChangeReceiver was being observed.

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
yadavshashankr: Removed the function and shifted the code in onActive()

Done

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
yadavshashankr: True

Done

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
yadavshashankr: It was from an earlier implementation, just pasted the code exactly as it is.

Rectified

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
yadavshashankr: Sorry for this. Currently I am desperately investing my thought process in understanding the architecture and that's why compromising it in these places. I'll try to remember to provide proper names in future PRs.

Done

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
yadavshashankr: Removed the function and shifted the code in onActive()

Done

app/src/main/java/com/portfolio/tasky/usecases/NetworkStatus.kt
yadavshashankr: I would request to keep this implementation as it is for now. I am gradually learning best practices,  and I really find this implementation easy to understand. I also assure that I will change the implementation to a better one in a future PR

app/src/main/java/com/portfolio/tasky/di/modules/AppModule.kt
```diff
@@ -0,0 +1,28 @@
...
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+class AppModule {
+
+    @Provides
+    fun providesContext(application: Application): Context {
+        return application.applicationContext
+    }
```
MKiperszmid, line 21: No need to do this. I'll explain in the next comment

app/src/main/java/com/portfolio/tasky/di/modules/AppModule.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Singleton
+    @Provides
+    fun provideNetworkChangeReceiver(context: Context): LiveData<NetworkStatus> {
```
MKiperszmid, line 25: Here you could simply add the ApplicationContext annotation, and it saves you to provide the context manually like you did before

Like this:
```
fun provideNetworkChangeReceiver(@ApplicationContext context: Context)
```

MKiperszmid (approved):
Just one small comment

app/src/main/java/com/portfolio/tasky/entry/viewModels/EntryViewModel.kt
```diff
@@ -0,0 +1,12 @@
...
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.ViewModel
+import com.portfolio.tasky.usecases.NetworkStatus
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+
+@HiltViewModel
+class EntryViewModel @Inject constructor(networkStatus: LiveData<NetworkStatus>) : ViewModel() {
+    val networkStatusObserver : LiveData<NetworkStatus> = networkStatus
+}
```
philipplackner, line 12: Here's my recommendation how I would handle observing network status while following clean architecture:
Listening to these status changes is part of the data layer, since you interact with an API (Android's NetworkManager API which is a system service). Therefore the NetworkObserver should go in your data package. However, since in clean architecture the presentation layer (where your ViewModel is in) isn't allowed to talk to data, what you need is an abstraction in domain. Here a sample how that could look like:
```
// In domain
interface NetworkObserver {
    val networkStatus: LiveData<NetworkStatus>
}

// In data
class NetworkObserverImpl(private val context: Context): NetworkObserver {
    override val networkObserver: LiveData<NetworkStatus> = NetworkChangedReceiver(context)
}

// In your VM
class EntryViewModel @Inject constructor(private val observer: NetworkObserver): ViewModel() { ... }

// In your AppModule.kt
@Provides
@Singleton
fun provideNetworkObserver(@ApplicationContext context: Context): NetworkObserver {
    return NetworkObserverImpl(context)
}
```

Why do we do all this abstraction stuff? When we test stuff in Android, there are 2 types of tests we normally have:
1. Local unit tests which run directly on the JVM and don't need an Android device. These are very fast, but only work if they don't use Android specific dependencies such as Context (which your NetworkChangedReceiver needs)
2. Instrumented tests which can use such Android dependencies, but need to run on an emulator which makes them very slow.

You want to test as much as possible with local unit tests, especially ViewModel logic. That's why we need to keep our ViewModels free from Android dependencies or classes which need Android dependencies. By creating an interface for our NetworkObserver which we pass to the ViewModel, we could create a fake version of it for testing, since it doesn't force us to pass the NetworkObserverImpl. We could instead pass a version which doesn't need the context and just simulates the network status changes. Quite complex topic, in the video portal you'll find a whole playlist on testing. If it's unclear, please ask in WhatsApp 👍

philipplackner (approved):

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
```diff
@@ -0,0 +1,67 @@
+package com.portfolio.tasky.usecases
```
philipplackner, line 1: This is not a use case, since it interacts with an API (the Android NetworkManager API). This should be in your data layer

philipplackner (approved):

app/src/main/java/com/portfolio/tasky/usecases/NetworkChangeReceiver.kt
yadavshashankr: Thanks for the info. Created a package data and shifted it there.

app/src/main/java/com/portfolio/tasky/di/modules/AppModule.kt
MKiperszmid: Interesting. Sounds good then, let's leave it as it is

PR #6, opened 2023-03-08
Feature : Retrofit addition, Login and Registration api calls.

yadavshashankr:


yadavshashankr:
Uploading WhatsApp Video 2023-03-09 at 02.08.20.mp4…

yadavshashankr:
https://user-images.githubusercontent.com/81215694/223844532-6924598b-5590-45b7-b227-db0c1cd7350f.mp4

yadavshashankr:
https://user-images.githubusercontent.com/81215694/223845102-c3483135-278c-46d9-a4ae-420ab98ff1a4.mp4

app/build.gradle
```diff
@@ -3,6 +3,7 @@ plugins {
     id 'org.jetbrains.kotlin.android'
     id 'kotlin-kapt'
     id 'com.google.dagger.hilt.android'
+    id 'com.google.android.libraries.mapsplatform.secrets-gradle-plugin'
```
MKiperszmid, line 6: What's the use of this plugin?

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -16,6 +16,8 @@ import com.portfolio.tasky.entry.fragments.LoginFragment
 import com.portfolio.tasky.entry.viewModels.EntryViewModel
 import com.portfolio.tasky.globals.Constants
 import com.portfolio.tasky.usecases.*
+import com.portfolio.tasky.usecases.domain.FragmentInflater
+import com.portfolio.tasky.usecases.domain.ToolbarHandler
```
MKiperszmid, line 20: Are you using these imports?

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
```diff
@@ -40,7 +44,17 @@ class RegistrationFragment : Fragment(), FragmentInflater by FragmentInflaterImp
...
-            startLoginFragment()
+            val name = viewBinding.etName.subLayout.etInput.text.toString()
+            val email = viewBinding.etEmail.subLayout.etInput.text.toString()
+            val password = viewBinding.etPassword.subLayout.etInput.text.toString()
+            viewModel.makeRegistrationCall(RegisterRequest(name, email, password)).observe(viewLifecycleOwner){
+                    it.let {  if(it == true){
+                        Toast.makeText(activity, "Success", Toast.LENGTH_SHORT).show()
+                    }else{
+                        Toast.makeText(activity, "Failure", Toast.LENGTH_SHORT).show()
+                    }
```
MKiperszmid, line 55: Here you could just
```
if(it == true){
    Toast.makeText(activity, "Success", Toast.LENGTH_SHORT).show()
}else{
    Toast.makeText(activity, "Failure", Toast.LENGTH_SHORT).show()
}
```

Avoid the unnecesary .let in this case

app/src/main/java/com/portfolio/tasky/entry/models/AuthenticationModel.kt
```diff
@@ -0,0 +1,9 @@
+package com.portfolio.tasky.entry.models
+
+import com.google.gson.annotations.Expose
+import com.google.gson.annotations.SerializedName
+
+data class AuthenticationRequest(@Expose @SerializedName("email")var email : String?, @Expose @SerializedName("password")var password : String?)
+
+data class AuthenticationResponse(@Expose @SerializedName("token")var token : String?, @Expose @SerializedName("userId")var userId : String?,
+                               @Expose @SerializedName("fullName")var fullName : String? )
```
MKiperszmid, line 9: I'm not a fan of having this in just one line
I rather have it like:

```
data class AuthenticationRequest(
    @Expose @SerializedName("email") var email : String?, 
    @Expose @SerializedName("password") var password : String?
)
```

It's a bit easier to read

MKiperszmid: Also, you shouldn't need to use @Expose

Since @Expose sets the serialization to true by default. Which if you don't add it, is the default value

So just having @SerializedName is more than enough

app/build.gradle
```diff
@@ -59,6 +60,13 @@ dependencies {
     implementation 'com.intuit.ssp:ssp-android:1.1.0'
     implementation 'com.intuit.sdp:sdp-android:1.1.0'
 
+    // Retrofit Networking
+    implementation 'com.squareup.retrofit2:retrofit:2.9.0'
+    implementation 'com.squareup.retrofit2:converter-gson:2.9.0'
```
MKiperszmid, line 65: I'd recommend to use anything but Gson for JSON parsing. 
Gson is a Java library and has its issues with things like nullability and default values in Kotlin (for example, it can assign null values to nullable fields). 

Checkout Moshi. Works really well with Retrofit ;)

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
```diff
@@ -0,0 +1,17 @@
...
+import retrofit2.http.Body
+import retrofit2.http.POST
+
+interface EntryApiCall {
+
+    @POST("/register")
+    fun registration(@Body registerModel: RegisterRequest) : Call<Void>
+
+    @POST("/login")
+    fun login(@Body authenticationModel: AuthenticationRequest) : Call<AuthenticationResponse>
```
MKiperszmid, line 16: You don't need to use "Call"
+ if you don't have suspend keyword, then you aren't using coroutines

So:
```
@POST("/register")
suspend fun registration(@Body registerModel: RegisterRequest) : ResponseBody // This can be used if you don't want a body, but want to know if 200, 404, etc

@POST("/login")
suspend fun login(@Body authenticationModel: AuthenticationRequest): AuthenticationResponse
```

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
```diff
@@ -0,0 +1,70 @@
...
+                }
+
+            }
+
+            override fun onFailure(call: Call<Void>, t: Throwable) {
+                taskyCallStatusCallback.onFailure(0, t.message.toString())
+                registrationLiveData.postValue(false)
+            }
+
+        })
```
MKiperszmid, line 42: If you remove Call + use coroutines, you can simply replace this with

```
try {
val result = entryApiCall.registration(registerModel)
// Do something with result
} catch(e: CancellationException) {
    throw e
} catch(e: Exception) {
// Do something here
}

```

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
```diff
@@ -0,0 +1,70 @@
...
+import retrofit2.Response
+import javax.inject.Inject
+
+class EntryRepository @Inject constructor(private val entryApiCall: EntryApiCall)  {
+    private val registrationLiveData = MutableLiveData<Boolean?>()
+    private val loginLiveData = MutableLiveData<AuthenticationResponse?>()
+
+    fun doRegistration(taskyCallStatusCallback: TaskyCallStatus, registerModel: RegisterRequest) : LiveData<Boolean?>{
+
+        taskyCallStatusCallback.onRequestCallStarted()
```
MKiperszmid, line 22: Since you'll be using coroutines, you won't need a callback like this one

You can just wait for the response, since there's no need for callbacks in coroutines

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
```diff
@@ -0,0 +1,70 @@
...
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+import javax.inject.Inject
+
+class EntryRepository @Inject constructor(private val entryApiCall: EntryApiCall)  {
+    private val registrationLiveData = MutableLiveData<Boolean?>()
+    private val loginLiveData = MutableLiveData<AuthenticationResponse?>()
+
+    fun doRegistration(taskyCallStatusCallback: TaskyCallStatus, registerModel: RegisterRequest) : LiveData<Boolean?>{
```
MKiperszmid, line 20: suspend fun doRegistration...

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
```diff
@@ -34,5 +40,8 @@ class LoginViewModel @Inject constructor(
     fun areFieldsValid() {
         mutableFieldsValid.value =  mutableEmail.value == true && mutablePassword.value == true
     }
+    fun makeLoginCall(authenticationModel: AuthenticationRequest) : LiveData<AuthenticationResponse?>{
+        return entryRepository.doLogin(taskyCallStatus, authenticationModel)
```
MKiperszmid, line 44: When using coroutines, you'll have to use 
```
viewModelScope.launch {
entryRepository.doLogin...
}
```

app/src/main/java/com/portfolio/tasky/networking/usecases/TaskyCallStatusImpl.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.portfolio.tasky.networking.usecases.domain.TaskyLoader
+import com.portfolio.tasky.networking.usecases.domain.TaskyCallStatus
+import javax.inject.Inject
+
+/**
+ * TaskStatusImpl will harbour the code for un-successful api calls and
+ * would present errors, messages in dialogs
+ * with respect to Response codes.
+ */
+class TaskyCallStatusImpl @Inject constructor(private val taskyLoader: TaskyLoader, private val context: Context) : TaskyCallStatus {
```
MKiperszmid, line 14: This whole class won't be needed with coroutines

app/src/main/java/com/portfolio/tasky/networking/usecases/TaskyLoaderImpl.kt
```diff
@@ -0,0 +1,19 @@
+package com.portfolio.tasky.networking.usecases
+
+import android.content.Context
+import android.widget.Toast
+import com.portfolio.tasky.networking.usecases.domain.TaskyLoader
+/**
+ * TaskyLoader will harbour the code for loading dialogs and would start or dismiss based on the value isLoading received in
+ */
+class TaskyLoaderImpl constructor(val context: Context) : TaskyLoader {
```
MKiperszmid, line 9: This one won't be needed either

app/src/main/java/com/portfolio/tasky/networking/usecases/domain/TaskyCallStatus.kt
```diff
@@ -0,0 +1,9 @@
+package com.portfolio.tasky.networking.usecases.domain
+
+interface TaskyCallStatus {
```
MKiperszmid, line 3: This one either

MKiperszmid (approved):

app/build.gradle
yadavshashankr: It helps import properties from local.properties in project. That is where I have kept my api key.

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
yadavshashankr: Done. Current implementation checks for a non-null value, if present then only Toast is called.

app/src/main/java/com/portfolio/tasky/entry/models/AuthenticationModel.kt
yadavshashankr: Single line  - Done

I tried keeping only @SerializedName but then it doesn't work. I have encountered such scenarios before where @Expose becomes necessary to add since it tells the retrofit that the variable can be used for api communication

app/build.gradle
yadavshashankr: I implemented Moshi but it asks me to writ custom adapters for every Data class I have. It doesn't do it by default.

I know it has benefits however I am confident with Gson

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
yadavshashankr: I would love to use co-routine however my current implementation is making it tough to implement it. If you can see that when a response is received, it is passed at multiple places. Kinda like the values and value received event is spreading at various implementations. 

Ass soon as the response is received the callback sends the result to TaskyStatus and from there a loading value is passed to TaskyLoader

This helps me to receive response at multiple places and I can utilise these responses as per convenience. 

e.g. Response is sent to multiple places like TaskyStatusImpl and LoginFragment

If I have a valid response code say 200 then I won't trigger anything in TaskyStatusImpl and the received response will be processed in LoginFragment.

However if I receive code 400 or 500 then TaskyStatusImpl will show up a dialog and the result sent to LoginFragment will be null and nothing will be processed in it.

If you see the TaskyLoader is completely controlled by TaskySatus and whenever an api call begins and ends it automatically shows start loading toast and completed loading toast. Instead of this toast I will have my custom progressbar which will enable disable from the call back of TaskyStatus. This also helps me show and hide loader and also react correctly to response codes  by just adding 3 lines at required places.

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
yadavshashankr: It will require some more code in order to make it work with current implementation

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
yadavshashankr: Current implementation requires this

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
yadavshashankr: Current implementation requires this

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
yadavshashankr: Current implementation not going with this. will find a better way to implement coroutine with existing functionality.

app/src/main/java/com/portfolio/tasky/networking/usecases/TaskyCallStatusImpl.kt
yadavshashankr: This is required since I want to handle error for api in dialogs over here.

app/src/main/java/com/portfolio/tasky/networking/usecases/domain/TaskyCallStatus.kt
yadavshashankr: This is required since I want to start stop loading UI over here.

app/build.gradle
yadavshashankr: I'll target a new PR specially for this ::thumbs up::

app/src/main/java/com/portfolio/tasky/entry/models/AuthenticationModel.kt
yadavshashankr: Once I implement Moshi, will get rid of this code.

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
yadavshashankr: Done

yadavshashankr: I have kept the call back but switched to coroutines.

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
yadavshashankr: Done

app/build.gradle
MKiperszmid: No need to make custom adapters. Instead of using @SerializedName , just use: @field:Json

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepository.kt
MKiperszmid: Awesome!

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
MKiperszmid: If you have all your api calls in a Repository, then the result of the api call will just be used in one place, and then the repository would be called wherever it's needed

But the important part, is that you could just have a try catch in the repository once, and it will do all the error handling there

app/src/main/java/com/portfolio/tasky/networking/usecases/TaskyCallStatusImpl.kt
MKiperszmid: Won't be needed when using Coroutines, since you can basically return a true/false (this is simplified) to the viewmodel, and then the viewmodel can show an error or success

And in your viewmodel, before calling the repository for login/register, you can set the isLoading to true, and on success or failure from repository, you can set the isLoading to false

So you do state managing inside the viewmodel, and api call inside the repository

Remember, the jobs are:
UI = Display the state, and modify the UI based on the state. Send user interactions to the viewmodel (Ex: User clicked a button)
ViewModel = Modify the state based on the user interactions or something like an api/db call from repository
Repository = Make api and db calls by accessing the retrofit api or room dao. This information will then be returned to the viewmodel

app/build.gradle
yadavshashankr: Tried this as well, still it gave adapter error everytime :(

Will try to do things again from scratch in the next PR

app/src/main/java/com/portfolio/tasky/networking/usecases/TaskyCallStatusImpl.kt
yadavshashankr: This explanation clears the picture 👍

yadavshashankr:
https://user-images.githubusercontent.com/81215694/224415719-e2c2ccce-21a2-451a-9afa-3f2007498a88.mp4

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val registrationLiveData = MutableLiveData<Boolean>()
+    private val loginLiveData = MutableLiveData<AuthenticationResponse?>()
+
+    private val coroutineExceptionHandler = CoroutineExceptionHandler{ _, throwable ->
+        throwable.printStackTrace()
+    }
+
+    override fun doRegistration(registerRequest: RegisterRequest): MutableLiveData<Boolean> {
+
+        CoroutineScope(Dispatchers.IO + coroutineExceptionHandler).launch {
```
yadavshashankr, line 26: Reason to call coroutine here is because I wanted to observe proper instance of LiveDatas in Login and Registration fragment.

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -67,7 +69,7 @@ class EntryActivity : AppCompatActivity(), FragmentInflater by FragmentInflaterI
         setTitle(getString(R.string.welcome_back))
 
         setFragmentManager(supportFragmentManager)
-        val loginFragment = LoginFragment.getInstance()
+        val loginFragment = RegistrationFragment.getInstance()
```
MKiperszmid, line 72: login fragment yet you get an instance of a RegistrationFragment?

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
```diff
@@ -40,7 +42,15 @@ class RegistrationFragment : Fragment(), FragmentInflater by FragmentInflaterImp
...
         viewBinding.btnReg.setOnClickListener {
-            startLoginFragment()
+            val name = viewBinding.etName.subLayout.etInput.text.toString()
+            val email = viewBinding.etEmail.subLayout.etInput.text.toString()
+            val password = viewBinding.etPassword.subLayout.etInput.text.toString()
+
+            viewModel.makeRegistrationCall(RegisterRequest(name, email, password)).observe(viewLifecycleOwner){
+                if(it){
+                    startLoginFragment()
+                }
```
MKiperszmid, line 52: If registration fails, we can show an error or something

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
```diff
@@ -64,6 +76,17 @@ class RegistrationFragment : Fragment(), FragmentInflater by FragmentInflaterImp
         viewModel.validateFields.observe(viewLifecycleOwner) {
             it.let {  viewBinding.btnReg.isEnabled = it == true }
         }
+
+        viewModel.networkObserver.observe(viewLifecycleOwner){
```
MKiperszmid, line 80: No need to have a network observer on the login/registration

You could just make the API call, and display the error

app/src/main/java/com/portfolio/tasky/entry/models/AuthenticationModel.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import com.google.gson.annotations.Expose
+import com.google.gson.annotations.SerializedName
+
+data class AuthenticationRequest(@Expose @SerializedName("email")var email : String?,
+                                 @Expose @SerializedName("password")var password : String?)
+
+data class AuthenticationResponse(@Expose @SerializedName("token")var token : String?,
+                                  @Expose @SerializedName("userId")var userId : String?,
+                                  @Expose @SerializedName("fullName")var fullName : String?)
```
MKiperszmid, line 11: Why are these nullable?

Request will always be non null since you'll create it based on user input

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
```diff
@@ -0,0 +1,17 @@
...
+import retrofit2.http.Body
+import retrofit2.http.POST
+
+interface EntryApiCall {
+
+    @POST("/register")
+    suspend fun registration(@Body registerModel: RegisterRequest) : Response<Void>
+
+    @POST("/login")
+    suspend fun login(@Body authenticationModel: AuthenticationRequest) : Response<AuthenticationResponse>
```
MKiperszmid, line 16: No need to wrap it in a "Response" class

You could simply return the object

Ex: On login: AuthenticationResponse
And on Registration:  ResponseBody (This is used instead of Void, so you can check if it returned a 200, 404, etc)

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
```diff
@@ -0,0 +1,59 @@
...
+
+class EntryRepositoryImpl@Inject constructor(private val entryApiCall: EntryApiCall, private val taskyCallStatus: TaskyCallStatus) : EntryRepository  {
+    private val registrationLiveData = MutableLiveData<Boolean>()
+    private val loginLiveData = MutableLiveData<AuthenticationResponse?>()
+
+    private val coroutineExceptionHandler = CoroutineExceptionHandler{ _, throwable ->
+        throwable.printStackTrace()
+    }
+
+    override fun doRegistration(registerRequest: RegisterRequest): MutableLiveData<Boolean> {
```
MKiperszmid, line 24: Why MutableLiveData?

Also, LiveData, Flows, etc is used when you'll get more than just 1 result
Here, you'll only get 1. Either a success with a body, or an error

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
```diff
@@ -0,0 +1,59 @@
...
+            }else{
+                val jObjError = response.errorBody()?.string()?.let { JSONObject(it) }
+                taskyCallStatus.onFailure(response.code(), jObjError?.getString("message") as String)
+                registrationLiveData.postValue(false)
+            }
+        }
+        return registrationLiveData
+    }
+
+    override fun doLogin(authenticationRequest: AuthenticationRequest) : LiveData<AuthenticationResponse?>{
```
MKiperszmid, line 42: @Same comment as above regarding live data


I'd recommend just returning something like a Kotlin Result that wraps the Authentication information


Also, you are returning a Data object back to the presentation
You shouldn't have that happening

You should have:

Data model, maps to a Domain model, and use that domain model in the presentation
Iif not, the presentation is knowing stuff from the Data layer. Which shouldn't be the case

app/src/main/java/com/portfolio/tasky/entry/viewModels/RegisterViewModel.kt
```diff
@@ -40,6 +48,10 @@ class RegisterViewModel  @Inject constructor(
         mutableName.value = nameValidation.isValidName(name)
     }
     fun areFieldsValid() {
-        mutableFieldsValid.value =  mutableEmail.value == true && mutablePassword.value == true && mutableName.value == true
+        mutableFieldsValid.value =  mutableEmail.value == true && mutablePassword.value == true && mutableName.value == true && networkObserver.value == NetworkStatus.Available
```
MKiperszmid, line 51: I recommend extracting this on a variable, since it will make it a bit more clearer

Ex:
```
val hasCompletedForm= mutableEmail.value == true && mutablePassword.value == true && mutableName.value == true
mutableFieldsValid.value = hasValidatedForm
```

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
yadavshashankr: On failure of registration we already show error from TaskyStatusImpl. You can check the video how it shows error on entering wrong login password, same happens for registration.  This is just a dummy functionality.

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
yadavshashankr: Instead of showing error from api call, I disable the buttons when network is not available.

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
yadavshashankr: At this stage this is making my implementation look simple. I can use the result as Response<> and process it in one place whether success or failure and return the data in case of success. Very easy as per my understanding. Will replace it with better version  in future.

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
yadavshashankr: because I want to update the LiveData who has this MutableLiveData reference.

Currently I need only true or false in terms of registration, thats why used a boolean liveData. On success or failure it will update the instance of the live data that is being observed on UI.

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
yadavshashankr: I need a little more understanding on how this will be implemented.

app/src/main/java/com/portfolio/tasky/entry/viewModels/RegisterViewModel.kt
yadavshashankr: True

Done

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
yadavshashankr: I just switch the fragment here for convenience.

Done.

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
yadavshashankr: Also since what I have implemented is a skeleton code to achieve the objective of successful login and retrieving access token. The idea is to not have token here, this is just a sample code which will change in future PR.

yadavshashankr: Also it will be great if you could direct me to some video achieving what you have explained here!!

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
MKiperszmid: Right,. But here you could just return a boolean (no need for live data), and change the livedata in the ViewModel based on the result

app/src/main/java/com/portfolio/tasky/entry/repositories/EntryRepositoryImpl.kt
MKiperszmid: I don't have a video, but I made an interview couple months ago where I applied something like it
https://github.com/MKiperszmid/StoreDiscount/blob/master/app/src/main/java/com/mkiperszmid/storediscount/core/data/ProductRepositoryImpl.kt#L17

In this case, I returned a List. But you could return the object

Also, there I applied what I mentioned about mapping objects from Data to Domain
Let me know if you need more explanation :D

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -42,7 +53,18 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl(), Te
 
         setObservers()
         viewBinding.btnLogin.setOnClickListener {
-            openRegistrationFragment()
+            val email = viewBinding.etEmail.subLayout.etInput.text.toString()
+            val password = viewBinding.etPassword.subLayout.etInput.text.toString()
+
+            lifecycleScope.launch(Dispatchers.IO + coroutineExceptionHandler){
+                viewModel.makeLoginCall(AuthenticationRequest(email, password))
```
MKiperszmid, line 60: rename the method from "makeLoginCall" to something like "login"
Makes a bit more sense + you don't mention it's API related

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
MKiperszmid: Sounds good

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt

PR #7, opened 2023-03-13
Login & Registration complete

yadavshashankr:
1. Fab implementation.

2. Navigation between Login and Registration fragment.

3. Implementation for setting/getting Access Token.


app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -51,21 +53,18 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl(), Te
...
-                }
-            }
-        }
+        viewBinding.btnLogin.setOnClickListener(this)
+        viewBinding.tvRegister.setOnClickListener(this)
+    }
+
+    private fun setToolbar() {
+        setFragmentManager(activity?.supportFragmentManager as FragmentManager)
+        (activity as EntryActivity).setTitle((activity as EntryActivity).getString(R.string.welcome_back))
```
philipplackner, line 67: I find this line hard to read due to all the conversions. You could make it a bit more readable like this:
```
val parentActivity = requireActivity() as EntryActivity
parentActivity.setTitle(requireContext().getString(R.string.welcome_back))
```

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -141,18 +140,30 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl(), Te
...
-     * Currently not called
-     */
+
+    private fun startLogin() {
+        val email = viewBinding.etEmail.subLayout.etInput.text.toString()
+        val password = viewBinding.etPassword.subLayout.etInput.text.toString()
+
+        lifecycleScope.launch(Dispatchers.IO + coroutineExceptionHandler){
+            viewModel.login(AuthenticationRequest(email, password))
+        }
```
philipplackner, line 150: I don't know how your `login()` function of the ViewModel looks like, but since you execute it in a suspend function, I strongly assume it's a suspend function. Never expose suspend functions from a ViewModel to the UI. If you press login here and then a config change happens (such as a screen rotation), your activity will be destroyed, `lifecycleScope` will be cancelled and logging in as well.
Instead, always launch coroutines in `viewModelScope` and update state afterwards which you observe in the UI 👍 

You very rarely really need `lifecycleScope` in Android.

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -141,18 +140,30 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl(), Te
...
+
+        viewModel.loginObserver.observe(viewLifecycleOwner){
+            if(it != null){
+                it.token?.let { it1 -> LoginUtils.setAccessToken(it1)}
+                Toast.makeText(activity, "Token ${it.token}", Toast.LENGTH_SHORT).show()
+            }
+        }
```
philipplackner, line 157: I'd avoid putting such data related stuff (like saving something in persistent storage) in the UI. Keep it as dumb as possible and only display something based on what you get from the ViewModel.

The following allows you to easily access user data from any layer in your app while being able to also easily swap out preferences implementations. The only thing needed for that would be to swap out the shared preferences object in the `UserPreferencesImpl` constructor with let's say a DataStore instance.

*Domain model*
```
data class AuthenticatedUser(
    val fullName: String,
    val email: String,
    val token: String
)
```

*Domain abstraction*
```
interface UserPreferences {
    fun saveAuthenticatedUser(user: AuthenticatedUser)
    fun getAuthenticatedUser(): AuthenticatedUser?
    fun clearPreferences()
}
```

*Data implementation*
```
class UserPreferencesImpl(
    private val sharedPref: SharedPreferences
): UserPreferences {
    override fun saveAuthenticatedUser(user: AuthenticatedUser) {
        // Save JSON of user in shared pref
    }

    override fun getAuthenticatedUser(): AuthenticatedUser? { ... }

    override fun clearPreferences() { ... }
}
```

In your Dagger module:
```
@Provides
@Singleton
fun provideUserPreferences(sharedPreferences: SharedPreferences): UserPreferences {
    // For sensitive user data like tokens, I'd use EncryptedSharedPreferences
    return UserPreferencesImpl(sharedPreferences)
}
```

Then, for example in any ViewModel:
```
@HiltViewModel
class MyViewModel @Inject constructor(
    private val userPreferences: UserPreferences
): ViewModel() {
    // Use userPreferences
}
```

Since your ViewModel knows if the login was successful or not, it can then use `userPreferences` to save the user data 👍 This way, you keep your preferences implementation data specific while being able to use it across the app and easily swap out how preferences are implemented.

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
```diff
@@ -49,21 +50,13 @@ class RegistrationFragment : Fragment(), FragmentInflater by FragmentInflaterImp
...
 
-            viewModel.registrationObserver.observe(viewLifecycleOwner){
-                if(it){
-                    startLoginFragment()
-                }
-            }
-        }
+    private fun setToolBar() {
+        setFragmentManager(activity?.supportFragmentManager as FragmentManager)
+        (activity as EntryActivity).setTitle((activity as EntryActivity).getString(R.string.create_your_account))
```
philipplackner, line 59: Same feedback as before

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
```diff
@@ -154,24 +147,37 @@ class RegistrationFragment : Fragment(), FragmentInflater by FragmentInflaterImp
...
-        setFragmentManager(activity?.supportFragmentManager as FragmentManager)
-        removeFragment(this)
+    private fun startRegistration() {
+        val name = viewBinding.etName.subLayout.etInput.text.toString()
+        val email = viewBinding.etEmail.subLayout.etInput.text.toString()
+        val password = viewBinding.etPassword.subLayout.etInput.text.toString()
+
+        lifecycleScope.launch(Dispatchers.IO + coroutineExceptionHandler) {
+            viewModel.registration(RegisterRequest(name, email, password))
+        }
```
philipplackner, line 163: Same feedback as before

app/src/main/java/com/portfolio/tasky/globals/LoginUtils.kt
```diff
@@ -0,0 +1,15 @@
...
+    companion object{
+        var token : String = ""
+        fun setAccessToken(token : String){
+            this.token = token
+        }
+
+        fun getAccessToken() : String{
+            return "Bearer $token"
+        }
+    }
```
philipplackner, line 14: Ah I now noticed that you don't have preferences yet. Not sure if using that was your plan or not, but to keep the user logged in, you'll need to store the token persistently. Using a companion object won't be enough.

philipplackner (approved):

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
yadavshashankr: Thanks for the information!

Done

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
yadavshashankr: Great info!

Done

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/entry/fragments/RegistrationFragment.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/globals/LoginUtils.kt
yadavshashankr: Removed this file

Done

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
```diff
@@ -1,17 +1,20 @@
...
 import com.portfolio.tasky.entry.models.AuthenticationResponse
 import com.portfolio.tasky.entry.models.RegisterRequest
 import retrofit2.Response
 import retrofit2.http.Body
+import retrofit2.http.Headers
 import retrofit2.http.POST
 
 interface EntryApiCall {
-
+    @Headers("x-api-key: ${BuildConfig.api_key}")
```
philipplackner, line 13: Since all of your routes require an API key it might make sense to use an interceptor for that which will add this header for all routes 👍

app/src/main/java/com/portfolio/tasky/entry/usecases/domain/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,26 @@
+package com.portfolio.tasky.entry.usecases.domain
```
philipplackner, line 1: This does neither belong in a use case package nor in domain, it's a data related thing. Probably just a mistake I assume :)

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
```diff
@@ -47,7 +56,11 @@ class LoginViewModel @Inject constructor(
...
         mutableFieldsValid.value = areFieldsValidated  && isNetworkAvailable
     }
-    suspend fun login(authenticationModel: AuthenticationRequest){
-        mutableLogin.postValue(entryRepository.doLogin(authenticationModel))
+    fun login(authenticationModel: AuthenticationRequest){
+        viewModelScope.launch(Dispatchers.IO + coroutineExceptionHandler) {
+            val authenticatedUser = entryRepository.doLogin(authenticationModel)
+            userPreferences.saveAuthenticatedUser(authenticatedUser as AuthenticationResponse)
+            mutableLogin.postValue(entryRepository.doLogin(authenticationModel))
+        }
```
philipplackner, line 64: Aren't you logging in twice here? Shouldn't it be `mutableLogin.postValue(authenticatedUser)` instead?

app/src/main/java/com/portfolio/tasky/entry/viewModels/RegisterViewModel.kt
```diff
@@ -55,7 +62,9 @@ class RegisterViewModel  @Inject constructor(
         val areFieldsValidated = mutableEmail.value == true && mutablePassword.value == true && mutableName.value == true
         mutableFieldsValid.value = areFieldsValidated  && isNetworkAvailable
     }
-    suspend fun registration(registerModel: RegisterRequest){
-        mutableRegistration.postValue(entryRepository.doRegistration(registerModel))
+    fun registration(registerModel: RegisterRequest){
```
philipplackner, line 65: Naming suggestion: `register()` since a function does something, so it should be a verb if it doesn't return anything 🤔

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptor.kt
```diff
@@ -1,19 +1,18 @@
...
 
 
-class RequestInterceptor @Inject constructor() :
-    Interceptor {
+class RequestInterceptor @Inject constructor(private val userPreferences: UserPreferences) : Interceptor {
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
         val original = chain.request()
         val requestBuilder = original.newBuilder()
-            .header("x-api-key", BuildConfig.api_key)
```
philipplackner, line 16: How comes you removed this?

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptor.kt
```diff
@@ -1,19 +1,18 @@
...
+            .header("Authorization", userPreferences.getAuthenticatedUser()?.token as String)
```
philipplackner, line 15: If the token doesn't exist, your app will crash here, instead I'd add a null check and only add the header if the token exists

philipplackner (approved):

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
yadavshashankr: I have kept this only for passing API_KEY

I am using a different interceptor for Access_Token

app/src/main/java/com/portfolio/tasky/entry/usecases/domain/UserPreferencesImpl.kt
yadavshashankr: okay..wasn't a mistake :D

still need a lot of info to understand this :)

app/src/main/java/com/portfolio/tasky/entry/viewModels/LoginViewModel.kt
yadavshashankr: Yes!!!
and this consumed all my time today :(

app/src/main/java/com/portfolio/tasky/entry/viewModels/RegisterViewModel.kt
yadavshashankr: Okay

Done

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptor.kt
yadavshashankr: Called it directly in the Api interface.

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptor.kt
yadavshashankr: I've handled this in UserPreference, where it will send an empty, non-null string to avoid that crash. It was getting difficult to check null and deal with non-null variables.

app/src/main/java/com/portfolio/tasky/networking/domain/RequestInterceptor.kt
```diff
@@ -0,0 +1,9 @@
+package com.portfolio.tasky.networking.domain
+
+import okhttp3.Interceptor
+import okhttp3.Response
+
+interface RequestInterceptor : Interceptor {
+
+    fun interceptRequest(chain : Interceptor.Chain) : Response
+}
```
MKiperszmid, line 9: The intercepting will only happen on Retrofit

So basically, on the data layer. Domain shouldn't know about it
So no need to have an interface for this. You can completely remove this file

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptorImpl.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.portfolio.tasky.entry.usecases.domain.UserPreferences
+import com.portfolio.tasky.networking.domain.RequestInterceptor
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.io.IOException
+import javax.inject.Inject
+
+
+class RequestInterceptorImpl @Inject constructor(private val userPreferences: UserPreferences) :
+    RequestInterceptor {
```
MKiperszmid, line 12: Just implement retrofit's Interceptor class

app/src/main/java/com/portfolio/tasky/entry/usecases/domain/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.content.Context
+import com.google.gson.Gson
+import com.portfolio.tasky.R
+import com.portfolio.tasky.entry.models.AuthenticationResponse
+import javax.inject.Inject
+
+class UserPreferencesImpl @Inject constructor(val context: Context) : UserPreferences {
+
+    private val sharedPreference by lazy { context.getSharedPreferences(context.getString(R.string.app_name), Context.MODE_PRIVATE)}
```
MKiperszmid, line 11: Since you are saving sensitive information, like a user token, I recommend using an EcryptedSharedPreference

app/src/main/java/com/portfolio/tasky/entry/usecases/domain/UserPreferencesImpl.kt
MKiperszmid: One rule of thumb

If it's android related, or it handles data, then it's 100% in the data layer

SharedPreferences is Android related
SharedPreferences ALSO handles data
So always in Data layer

app/src/main/java/com/portfolio/tasky/entry/usecases/domain/EmailPatternValidator.kt
```diff
@@ -1,4 +1,4 @@
-package com.portfolio.tasky.entry.usecases
+package com.portfolio.tasky.entry.usecases.domain
```
MKiperszmid, line 1: Usually domain has usecases, not the other way around

So the package should be:
entry -> Domain -> Usecases

app/src/main/java/com/portfolio/tasky/entry/network/EntryApiCall.kt
MKiperszmid: The API KEY will also be used in all the requests
So you could have another Interceptor, just for the API Key

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/networking/domain/RequestInterceptor.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptorImpl.kt
yadavshashankr: Done

app/src/main/java/com/portfolio/tasky/entry/usecases/domain/UserPreferencesImpl.kt
yadavshashankr: for current implementation I'll be keeping this and in a future PR I'll be storing access_token in keystore 👍

app/src/main/java/com/portfolio/tasky/data/RequestInterceptorApiKey.kt
```diff
@@ -1,14 +1,13 @@
-package com.portfolio.tasky.networking
+package com.portfolio.tasky.data
 
 import com.portfolio.tasky.BuildConfig
 import okhttp3.Interceptor
 import okhttp3.Response
 import java.io.IOException
 import javax.inject.Inject
 
+class RequestInterceptorApiKey @Inject constructor() : Interceptor{
```
MKiperszmid, line 9: This is just my personal preference, but I like to have the name of the interface at the end

Ex:
RequestApiKeyInterceptor

Or better yet
ApiKeyInterceptor

app/src/main/java/com/portfolio/tasky/entry/EntryActivity.kt
```diff
@@ -65,15 +82,20 @@ class EntryActivity : AppCompatActivity(), FragmentInflater by FragmentInflaterI
...
-        val fragment = RegistrationFragment.getInstance()
         inflateFragment(fragment, R.id.fragment_container)
     }
 
     fun setTitle(text: String) {
         setToolBarText(viewBinding.toolbar, text)
     }
+
+    override fun onClick(view: View?) {
+        if(view == viewBinding.fab && viewBinding.fab.tag == "viewTag"){
```
MKiperszmid, line 97: I'm not a fan of this
Maybe have a callback interface or something 

Or in the worst case, make it a constant

The reason why I don't like this, is that this activity needs to know about a string that's on a fragment

app/src/main/java/com/portfolio/tasky/entry/data/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+        val jsonAuthenticatedUser = Gson().toJson(authenticatedUser)
+        sharedPreference.edit().putString("authUser", jsonAuthenticatedUser).apply()
+    }
+
+    override fun getAuthenticatedUser(): AuthenticationResponse? {
+        val authenticatedUserJson = sharedPreference.getString("authUser", "")
+        return if(!authenticatedUserJson.equals("") && !authenticatedUserJson.equals("null")){
+            Gson().fromJson(authenticatedUserJson, AuthenticationResponse::class.java)
+        }else{
+            AuthenticationResponse("","","")
```
MKiperszmid, line 23: You can return null here. Instead of an empty AuthenticationResponse

app/src/main/java/com/portfolio/tasky/entry/data/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.google.gson.Gson
+import com.portfolio.tasky.R
+import com.portfolio.tasky.entry.models.AuthenticationResponse
+import javax.inject.Inject
+
+class UserPreferencesImpl @Inject constructor(val context: Context) : UserPreferences {
+
+    private val sharedPreference by lazy { context.getSharedPreferences(context.getString(R.string.app_name), Context.MODE_PRIVATE)}
+
+    override fun saveAuthenticatedUser(authenticatedUser: AuthenticationResponse?) {
```
MKiperszmid, line 13: Make it non nullable

What would be the use of saving a null? None (or at least, I can't think of one)
So no need to do it

app/src/main/java/com/portfolio/tasky/entry/fragments/LoginFragment.kt
```diff
@@ -141,18 +131,25 @@ class LoginFragment : Fragment(), FragmentInflater by FragmentInflaterImpl(), Te
...
 
+    private fun startLogin() {
+        val email = viewBinding.etEmail.subLayout.etInput.text.toString()
+        val password = viewBinding.etPassword.subLayout.etInput.text.toString()
+
+        viewModel.login(AuthenticationRequest(email, password))
+
+        viewModel.loginObserver.observe(viewLifecycleOwner){
+            if(it != null){
+                Toast.makeText(activity, "Token ${it.token}", Toast.LENGTH_SHORT).show()
```
MKiperszmid, line 143: This is temporary right?

MKiperszmid (approved):

app/src/main/java/com/portfolio/tasky/entry/usecases/domain/UserPreferencesImpl.kt
philipplackner: Yeah when I learnt about this, I was also often confused because with data we often only associate things like a database or remote API. But, in the end we mean anything "external", so something outside of our app. This can be a remote backend, it can be a local database, it can be saving into a file, but also things like Android specific APIs, such as `NotificationManager`. In that case, it's not your app that shows the notification, instead it's a service from the OS which your app talks to, but still considered external. 

So in your domain layer you shouldn't have any imports that contain android or a 3rd party library.

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptor.kt
philipplackner: Here you'd only need to add this single line to make sure all your requests contain the key. In your API interface you'll need to add this for every single route

app/src/main/java/com/portfolio/tasky/networking/RequestInterceptor.kt
philipplackner: But `userPreferences.getAuthenticatedUser()` can apparently return `null` since you have a `?`-check there. So it seems like your prefs are sending a null `AuthenticatedUser` back if there is none (which is correct behavior)

app/src/main/java/com/portfolio/tasky/data/RequestInterceptorApiKey.kt
philipplackner: I agree 👍

app/src/main/java/com/portfolio/tasky/entry/data/UserPreferencesImpl.kt
philipplackner: Yes, don't be afraid to use null values. They clearly indicate that a value doesn't exist which is their job. With this value, you'd assume you actually got a response, but it's empty which will lead to bugs and unnecessary debugging

app/src/main/java/com/portfolio/tasky/entry/data/UserPreferencesImpl.kt
philipplackner: Agreed

